PR #3, opened 2019-07-17
Completed Problem 1

gauravj09:


problem1.cpp
```diff
@@ -0,0 +1,42 @@
...
+                    A[++n2] = 2; A[++n1] = 1;
+                }
+                else if (A[i] == 2) 
+                {
+                    A[++n2] = 2;
+                }
+            }
+        }
+};
+
```
ashmeetdeol19: Thats also a great solution Gaurav 👍 
Well Done!!

PR #5, opened 2019-07-17
Please review: Completed problems for day 12 Two pointer-1 ContainerWithMostWater, S…

MahiteeS:
…ortColors, ThreeSum

### DIFF
--- a/ContainerWithMostWater.java
+++ b/ContainerWithMostWater.java
@@ -1,0 +1,50 @@
+package day12_TwoPointers1;
+
+/*
+	Idea : 1. Take two pointers, one at start and one at end.
+		   2. Check for minimum height among both of them and find maxArea.
+		   3. To find the max area, shift pointer to left and right according to max-height and keep on updating the maxArea.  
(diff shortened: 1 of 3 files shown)


ContainerWithMostWater.java
```diff
@@ -0,0 +1,50 @@
+package day12_TwoPointers1;
+
+/*
+	Idea : 1. Take two pointers, one at start and one at end.
```
ashmeetdeol19: Thats a great explanation 💯

SortColours.java
```diff
@@ -0,0 +1,62 @@
+package day12_TwoPointers1;
+
+/*     Standard counting sort approach
+Idea : 1. Make a new array of size 3, and increase the count of index 0, 1, 2 according to 
+		  element values in nums array
+	   2. Place numbers 0, 1, 2 in nums array according to the values of temp array indexes
```
ashmeetdeol19: Thats a good solution but we were given a note that not to use counting sort.

PR #6, opened 2019-07-18
Completed for problem1 using the two pointers approach.

lavanganji:


Problem1.java
```diff
@@ -0,0 +1,98 @@
...
+		int[] arr = {2,0,2,1,1,0};
+		s.sortColors(arr);
+		
+		TwoPointerSolution s2 = new TwoPointerSolution();
+		int[] arr2 = {2,0,2,1,1,0};
+		s2.sortColors(arr2);
+
+	}
+
+}
```
ashmeetdeol19: Its good to see that you have tried both the approaches 👍

Problem2.java
```diff
@@ -0,0 +1,49 @@
...
+                else
+                    r--;
+            }
+            
+        }
+        return res;
+	
+        
+    }
+}
```
ashmeetdeol19: Thats a good Solution 👍

PR #7, opened 2019-07-18
 Problem1 and Problem2 - Please review

hrshah5:


Problem2.java
```diff
@@ -0,0 +1,72 @@
...
+                            result.add(new ArrayList(temp));
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        return result;
+        
+    }
```
ashmeetdeol19: In such problems brute force solution givesbTLE.

PR #14, opened 2019-08-14
Finished Problem 1&2: Please review

muditabysani:


Problem2.py
```diff
@@ -0,0 +1,21 @@
+class Problem2:
+	def ThreeSum(self, arr):
+		# Time complexity : O(n^3) where n is the number of elements in the array
+		# Space complexity : result array. No extra space being used
+		# Finding every 3 elements combination from the array and adding it to the resultant array only if the sum is equal to zero
+		# But this solution has duplicates in it.
```
kaurramanpreet: can you explain why this solution has duplicates?

muditabysani: After I sort the each 3 item list and checking for its existence in the resultant array, there were no duplicates in it. The duplicates are because of the repetition of elements in the array

PR #22, opened 2019-09-05
Please review.

Jaysparkexel:
> Please check **most recent commit**.

### DIFF
--- a/Problem2.py
+++ b/Problem2.py
@@ -1,3 +1,12 @@
+# Time Complexity : O(n^2) 
+# Space Complexity : O(n) 
+# Did this code successfully run on Leetcode : Yes
+# Three line explanation of solution in plain english:
+# - Sort the nums. Initialize answer as set.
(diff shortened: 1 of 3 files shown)


Problem2.py
```diff
@@ -0,0 +1,44 @@
...
+        :type nums: List[int]
+        :rtype: List[List[int]]
+        """
+#       Starting with answer as set to store only unique triplet
+        res = set()
+        
+#       If length is less than 3 than we can't create triplet
+        if len(nums) < 3:
+            return res
+#       Sorting the number because, sorting time complexity: O(n) and algorithm takes more than that time complexity.
```
kaurramanpreet: sorting time complexity is o(nlogn) not o(n)

Jaysparkexel: Got it @kaurramanpreet . Wrote it by mistake. Can you also review Problem 3 that I added recently?

PR #48, opened 2019-12-04
Update maxwater.java

sakshi1705:


maxwater.java
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Solution {
+    public int maxArea(int[] height) {
+        int max=0;
+        for(int i=0;i<height.length;i++)
+                for(int j=i+1;j<height.length;j++)
+                    max= Math.max(max, Math.min(height[i],height[j])*(j-i));
+            return max;
+
```
neha79091: Try to optimize it

sortflag.java
```diff
@@ -0,0 +1,26 @@
+class Solution {
+  
+  public void sortColors(int[] nums) {
```
neha79091: Good. Write complexity also.

PR #51, opened 2019-12-05
Completed Two-pointers-1

SweetySojrani:


Problem2.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity - O(n^2) where n is the number of elements in the array
+//Space Complexity - O(1)
```
neha79091: How is time complexity O(n^2)

SweetySojrani: For each i th element in the array,  it is traversing from i+1 till the end to get unique possible triplets.

neha79091: You are traversing the elements once. Recall the algorithm.

neha79091: The complexity given by you is the correct one.

PR #52, opened 2019-12-05
Adding 2 pointer changes

dasgprat:
Please review the PR and also let me know in case of any issue or concerns.

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,55 @@
+
+/**
+Leet Code Submitted : YES
+Space Complexity : O(1)
+Time Complexity : O(N^2)
+
+The idea is to iterate each position in the array after sorting and check if the sum goes overflow or underflow and accordingly increases/decreases pointer.
+**/
+
+
+class Solution {
+    public List<List<Integer>> threeSum(int[] nums) {
+        
+        List<List<Integer>> out = new ArrayList<>();
+        
+        if(nums == null || nums.length < 1)
+            return out;
+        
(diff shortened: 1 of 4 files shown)


Problem2.java
```diff
@@ -0,0 +1,55 @@
+
+/**
+Leet Code Submitted : YES
+Space Complexity : O(1)
+Time Complexity : O(N^2)
+
```
neha79091: How is your time complexity O(n^2)

Problem3.java
```diff
@@ -0,0 +1,39 @@
...
+Time Complexity : O(N)
+
+The intuition is to check current max and increment based pointer (high/low) based on the minimum height of the bar and accordingly increase high pointer and low pointer. Finally, returing max of the area 
+**/
+
+
+
+class Solution {
+    public int maxArea(int[] height) {
+
```
neha79091: You changed the return type of method(it is void).

Problem2.java
dasgprat: Yeah every element is checked only once. So, it should be O(N)

neha79091: it will be O(nlogn).You are sorting also, Take care while calculating complexity by algorithm, code will confuse you

super30admin: The time complexity given by you is the correct one. I misunderstood  it a little.

neha79091: Prateek, the complexity given by you is the correct one. I misunderstood it.

PR #54, opened 2019-12-05
Add files via upload

gunjanbatragandhi:
Please review

### DIFF
--- a/Problem#33.py
+++ b/Problem#33.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Wed Dec  4 19:29:56 2019
+
+@author: gunjan
+"""
+
+
+##################################
+###########Method1:
+#################################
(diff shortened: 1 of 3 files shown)


Problem#33.py
```diff
@@ -0,0 +1,56 @@
...
+#swap function
+def swap(x,y):
+    temp=x
+    x=y
+    y=temp
+    return(x,y)
+
+#solution function
+def solution(nums):
+
```
neha79091: There were two arguements

Problem#34.py
```diff
@@ -0,0 +1,78 @@
...
+                pt1+=1
+                pt2-=1
+            
+    return answer
+
+solution([-1,0,1,2,-1,-4])
+
+#Time Complexity=O(n^2) - for loop and while loop
+#Space Complexity=O(1)
+
```
neha79091: Think about complexity again.

PR #55, opened 2019-12-06
Complete all the problems - Please Review

devashishnyati:


3Sum.py
```diff
@@ -0,0 +1,36 @@
+# Time Complexity : O(n**2)
+# Space Complexity : O(1)
```
neha79091: Do you mean O(n^2)?

devashishnyati: @neha79091 yes

neha79091: think about the complexity again using algorithm

devashishnyati: @neha79091 I still think the time complexity is O(n^2)

jaspindersingh83: Devashish you are right...its an n2 solution.
@neha79091 Kindly consult me if you are confused about some solution given by a student

super30admin: Yes, your answer is correct, I misunderstood .

neha79091: yup, I misunderstood it a little.

PR #61, opened 2020-01-08
Two Pointer - 1

BhushanVasisht:


super30admin:
Good job Bhushan 
We will be looking at more optimal solutions in the class

PR #64, opened 2020-01-09
two pointers1

juhipathak:
please review it

### DIFF
--- a/3sum.java
+++ b/3sum.java
@@ -1,0 +1,40 @@
+// space complexity : O(1)
+// time complexity : O(n2)
+class Solution {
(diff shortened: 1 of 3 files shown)


3sum.java
```diff
@@ -0,0 +1,40 @@
+// space complexity : O(1)
+// time complexity : O(n2)
+class Solution {
+    public List<List<Integer>> threeSum(int[] nums) {
+        List<List<Integer>> list=new ArrayList<>();
+        if(nums.length<3 || nums==null) return list;
```
rajatkumar01: nums == null will be first condition as if nums is null  then nums.length will fail

PR #67, opened 2020-01-09
all problems complete as per class

nikrasiya:


15_3sum.py
```diff
@@ -0,0 +1,106 @@
+from typing import List
+
+class Solution:
+    def threeSum(self, nums: List[int]) -> List[List[int]]:
+        """
+                https://leetcode.com/problems/3sum/
+                // Time Complexity : O(nlogn + n^2)
+                // Space Complexity : O(n)
+                            Sorting requires O(n) space
```
super30admin: You are right although we are not sure what is the sorting algo used under the hood.  Keep it generalized

15_3sum.py
```diff
@@ -0,0 +1,106 @@
+from typing import List
+
+class Solution:
+    def threeSum(self, nums: List[int]) -> List[List[int]]:
+        """
+                https://leetcode.com/problems/3sum/
+                // Time Complexity : O(nlogn + n^2)
```
super30admin: Ignore n log n

nikrasiya: Sure

PR #69, opened 2020-01-10
Two Pointers-1 problem 2

nehabindle:
Please review my solution for 2 sum problem.
Space and time complexity is O(N)
Ran successfully on leetcode as well.


super30admin:
Hi Neha, the 3 sum problem is correct. I could not see the code for 2 sum

PR #84, opened 2020-02-20
Completed problem 1 and 2

PratikshaPP:
Sort colors - Couldn't figure out a one pass solution so did it using counting approach
3Sum - Used two pointers
Container with most water - Not done


Lohith-Aswathappa:
Good Job Prathiksha!

Try the other problem as well

PR #86, opened 2020-02-20
Two pointer solution

ptndoss:


BhushanVasisht:
Yes. These are correct.

PR #100, opened 2020-03-07
solved all three

sivasaimupparaju:
please review


BhushanVasisht:
Please give the order of runtime so that u understand which term is affecting the runtime and space..Dont keep it generic to linear

PR #114, opened 2020-04-03
Problem-2 completed

schakrad:


devashishnyati:
Good job Sindhu! Try to optimize it on space.
Submit the remaining problems also.

PR #117, opened 2020-04-05
adds the codes for two pointer1 problems

shahakshay11:


devashishnyati:
Good job Akshay!
Nice work with the explanations!

PR #133, opened 2020-05-08
Completed Design

JacobStephen9999:


3Sum.py
```diff
@@ -0,0 +1,34 @@
+#Time Complexity: O(NlogN) for sorting and O(N*N) where N is inputs
```
iivaishnavii, line 1: So asymptotically, it become n2.

iivaishnavii:
Great job, Jacob!! Keep it up..

PR #138, opened 2020-05-08
Please review these problems 3sum, sortColors, waterContainer

albertcoder:


3sum.java
```diff
@@ -0,0 +1,30 @@
+//time complexity O(n)
```
super30admin, line 1: O(n2) is correct time complexity

super30admin:
Good Job. Please review inline comment

3sum.java
albertcoder: Got it! It makes sense since in worst case scenario, the " i " pointer will iterate over the whole array (until nums.length -2) which is O(n )and the low and high pointer inside contribute to nested O(n) time complexity which makes it O(n^2) time complexity. Thanks for spotting that out.

super30admin: Yuup

PR #145, opened 2020-06-09
Completed SortColor and Container Problem-Please Check

DineshNadar95:


3Sum.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity :O(nlogn +n^2)
```
AbhiniveshP, line 1: There is no need of the term NlogN

AbhiniveshP:
Nice work Dinesh. Its good that you've submitted brute force approaches as well.

PR #169, opened 2020-06-11
completed Two-Pointers-1: Please review

nileshdev996:


Lohith-Aswathappa:
Nilesh, Good work with brute force and optimized solutions.

PR #170, opened 2020-06-11
Two Pointers Problem. Please Review

soumitra9:


Lohith-Aswathappa:
Soumitra, Good work with both the brute force and optimized solutions.

PR #184, opened 2020-06-15
SortColors.py

shivaallani7:


Lohith-Aswathappa:
Nice try, Shiva!
Try the other problems as well.

Lohith-Aswathappa:
Nice try, Shiva!
Try the other problems as well.

PR #229, opened 2020-09-03
Completed Two Pointers-1 problems

sindhusha-parimi:


khushal1198 (commented):
Very good job Sindhusha!

khushal1198:
Very good job Sindhusha!

PR #235, opened 2020-09-03
2 pointer set 1

sidhs5695:


khushal1198 (commented):
Good job Siddharth. Your approach seems to be fine!

khushal1198:
Good job Siddharth. Your approach seems to be fine!

PR #272, opened 2020-10-01
Completed 1&2 partially. 3 completed.

koshyanand:


Problem2.py
```diff
@@ -0,0 +1,22 @@
+# Time Complexity: O(N^2)
+# Space Complexity: O(N)
+# Did not pass leetcode. Need to optimize further to handle duplicates
+
```
Lohith-Aswathappa, line 4: Check the logic again after class. Let me know if you have doubt

Lohith-Aswathappa (commented):
Nice try, Anand!
Try the other problems as well.

Lohith-Aswathappa:
Nice try, Anand!
Try the other problems as well.

PR #279, opened 2020-10-01
Two Pointers-1 completed, Please verify

Sowmya246:


sum3.py
```diff
@@ -0,0 +1,30 @@
+// Time Complexity : O(n)
```
Lohith-Aswathappa, line 1: check the time complexity

Lohith-Aswathappa (commented):
Good work, Sowmya!
Check the inline comments

Lohith-Aswathappa:
Good work, Sowmya!
Check the inline comments

PR #280, opened 2020-10-01
init two pointer 1

amitsingh5949:


ramya-sruthi-d (commented):
Nice job, done very well with explaining your approach.

ramya-sruthi-d:
Nice job, done very well with explaining your approach.

PR #286, opened 2020-10-04
Problems 3 and 1

DhruvJawalkar:


SortColors.java
```diff
@@ -0,0 +1,47 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : failed for a case when I had while(curr<right) as the condition
```
kannan-mrugapphan: Always think whether a decision on right has been made or not, that will lead you to breaking condition

kannan-mrugapphan (commented):
Nice Work

kannan-mrugapphan:
Nice Work

PR #291, opened 2020-10-28
Two-Pointer-1 Problems - Please Review

jayapratap:


Problem 1.java
```diff
@@ -0,0 +1,22 @@
+// Problem 1: 
+
+// Time Complexity : O(n^2)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I believe it can be further optimized
```
kannan-mrugapphan, line 6: normal sorting itself takes only nlogn. Try the optimized approach to be discussed in class

Problem 2.java
```diff
@@ -0,0 +1,24 @@
+// Problem 2: 
+
+// Time Complexity : O(n^2)
+// Space Complexity : O(1)
```
kannan-mrugapphan, line 4: Try to optimize this

kannan-mrugapphan (commented):
Good Job with the brute force solutions.

kannan-mrugapphan:
Good Job with the brute force solutions.

PR #320, opened 2020-11-13
completed please review

Mohit0993:


problem2.py
```diff
@@ -0,0 +1,36 @@
+#timecomplexityO(n log n)
```
samarthjj, line 1: O(n^2) in the worst case.

samarthjj:
Good work Mohit, Please review the inline comments.

samarthjj:
Good work Mohit, Please review the inline comments.

PR #338, opened 2020-12-08
Two pointer 1 Please review it 

souravit:


keshav-Gupta-IS (commented):
Good Job, Sourav!! But please explain your approach for every question just not the one on top.

keshav-Gupta-IS:
Good Job, Sourav!! But please explain your approach for every question just not the one on top.

PR #360, opened 2021-01-08
sort color and water container solns

prav23:


maithrreye:
Good Job Praveen !

maithrreye:
Good Job Praveen !

PR #369, opened 2021-01-08
completed two pointers 1

Utsav-tech:


maithrreye:
Good work Utsav! Please add comments , TC and SC.

maithrreye:
Good work Utsav! Please add comments , TC and SC.

PR #385, opened 2021-01-11
Please review Two-Pointers1

bahaarsadhnani:


ramya-sruthi-d (commented):
The time and space complexity for threesum problem is wrong. You are sorting the nums so tc would be O(nlogn) and apart from the result you have created one more list so need to include it in the space complexity. The rest of the problems are fine.

ramya-sruthi-d:
The time and space complexity for threesum problem is wrong. You are sorting the nums so tc would be O(nlogn) and apart from the result you have created one more list so need to include it in the space complexity. The rest of the problems are fine.

PR #386, opened 2021-01-11
Two-Pointers-1 please review

lokeshsakamoori:


ramya-sruthi-d (commented):
Well done with code, explain your approach as well.

ramya-sruthi-d:
Well done with code, explain your approach as well.

PR #393, opened 2021-01-15
Submitting 2 pointers

yahska-ednihs:


samarthjj:
Good work Akshay, Please work out the optimized solution of the Three Sum problem as well.

samarthjj:
Good work Akshay, Please work out the optimized solution of the Three Sum problem as well.

PR #396, opened 2021-01-15
Two-Pointers-1 completed

dhvanipatel27:


ramya-sruthi-d (commented):
Good work on all the problems, but for the second problem you are using an additional list li, so consider it in the space complexity.

ramya-sruthi-d:
Good work on all the problems, but for the second problem you are using an additional list li, so consider it in the space complexity.

ramya-sruthi-d:
Ignore the previous comment about space complexity, everything is good.

PR #416, opened 2021-02-04
two  pointer 1 submission. Please review

urjagparekh:


3sum.java
```diff
@@ -0,0 +1,30 @@
+class Solution {
+    public List<List<Integer>> threeSum(int[] nums) {
+        
+        Arrays.sort(nums);
```
yahska-ednihs, line 4: Good Job! Please add comments and the sort colors problem.

ContainerWithMostWater.java
```diff
@@ -0,0 +1,33 @@
+class Solution {
+    
+    
+    public int maxArea(int[] height) {
+        
+        int x = 0;
```
yahska-ednihs, line 6: Good Job! Please add comments and the sort colors problem.

PR #491, opened 2021-05-09
two pointer

sunilhector:


ramya-sruthi-d (commented):
Great job with code, please explain your approach as well.

ramya-sruthi-d:
Great job with code, please explain your approach as well.

PR #518, opened 2021-06-10
Completed 2 pointers1, please review

SheetalMore9878:
Working on 3 sum problem, submitting first two


Rooppesh:
Good Job! Please add the TC and SC!

Rooppesh:
Good Job! Please add the TC and SC!

PR #558, opened 2021-06-22
Completed Two-Pointers-1 (Please Review)

raviKsah:


ramya-sruthi-d (commented):
Good work with code, mention the approach as well in the comments.

ramya-sruthi-d:
Good work with code, mention the approach as well in the comments.

PR #571, opened 2021-07-08
Please Review - Completed Sprint Two Pointers-1

Mangesh-Ashtankar:


Rooppesh:
Good Job Mangesh! Please add the TC and SC as well!

Rooppesh:
Good Job Mangesh! Please add the TC and SC as well!

PR #586, opened 2021-07-14
completed-two-pointer-1-please review

HarishvarKS:


sridhar-anish (commented):
Good Job, make sure to write TC and SC

sridhar-anish:
Good Job, make sure to write TC and SC

PR #612, opened 2021-08-20
Two-Pointers-1, Pls review

ysrinu1:


Rooppesh:
Good Job! Add the TC, SC and approach used!

Rooppesh:
Good Job! Add the TC, SC and approach used!

PR #957, opened 2022-07-08
Completed 2 pointers

aneesh2411:


daredevilteja:
Please add TC and SC before raising a PR.

PR #996, opened 2022-08-04
Completed Two-Pointer-1  Both 1  and 2 Problems

javeedsanganakal:


ThreeSum.java
```diff
@@ -0,0 +1,103 @@
+
+// Time Complexity : O(n^2logn)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode :
+
+
+//******
+// Any problem you faced while coding this : Can you  please explain time complex and space Complex for the below solutions
```
jeelpatel18, line 8: For sorting it will take n*logn + nested loop(for and while) which is doing in N^2 ---> Total TC : N^2 + N*LogN

jeelpatel18: If we ignore TC for sorting, then it will be only N^2

ThreeSum.java
```diff
@@ -0,0 +1,103 @@
+
+// Time Complexity : O(n^2logn)
```
jeelpatel18: Review the comments below

ThreeSum.java
```diff
@@ -0,0 +1,103 @@
...
+        //loop till 3rd num from last
+        for(int i=0; i < n-2; i++){
+            
+            // if
+            if(i!=0 && nums[i] == nums[i-1]) continue;
+            
+            int low = i+1;
+            int high = n-1;
+            
+            while(low<high){
```
jeelpatel18, line 36: When low and high is too far for example, then we can say this loop will take O(N) time.

ThreeSum.java
```diff
@@ -0,0 +1,103 @@
...
+                }
+            }
+        }
+        
+        return result; 
+    }
+}
+
+
+// Time Complexity : O(n^3logn)
```
jeelpatel18: N^3 + N*LogN you can consider the TC as previous

jeelpatel18 (commented):
Good job. Submit 3rd problem too and let me know if you have confusion in TC in given problem.

jeelpatel18:
Good job. Submit 3rd problem too and let me know if you have confusion in TC in given problem.

PR #999, opened 2022-08-06
Two Pointer -1 Completed Please Review

AvishKadakia:


Problem-2.py
```diff
@@ -0,0 +1,41 @@
+#  15. 3Sum
+'''
+Leetcode all test cases passed: Yes
+class Solution:
+    def threeSum(self, nums: List[int]) -> List[List[int]]:
+        n is the length of nums
+        Space Complexity: O(1)
```
daredevilteja, line 7: SC depends on the implementation of sorting algorithm. It can be anywhere from O(log n) to O(n)

daredevilteja:
Good Work! Keep it up

PR #1040, opened 2022-09-22
two pointers-1 done

charan005-dev:


pvk3081 (commented):
Good Job Charan

pvk3081:
Good Job Charan

PR #1053, opened 2022-09-28
Completed Two Pointers 1. Please Review.

lakshaynanda:


3sum.java
```diff
@@ -0,0 +1,40 @@
+// TC: O(n^2)
+// SC: O(1)
```
daredevilteja, line 2: SC is O(logn) to O(n) due to sorting

daredevilteja:
Good Work! Keep it up

PR #1079, opened 2022-10-13
Completed Sprint Two-Pointers-1 : Please Review

jaszaveri:


Question2.py
```diff
@@ -0,0 +1,26 @@
+#Time Complexity : O(n*n)
+#Space Complexity : O(1)
```
daredevilteja, line 2: SC is O(n) because of sorting

daredevilteja:
Good Work! Keep it up

PR #1082, opened 2022-10-19
Completed two pointer

gourishpisal08:


3sum.py
```diff
@@ -0,0 +1,27 @@
+#TC=O(n2)
+#SC=O(1)
```
daredevilteja, line 2: SC is O(n) due to sorting

daredevilteja:
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

PR #1088, opened 2022-10-23
Two Pointer 1 all 3 questions are solved. Please review it.

dharatipatel123:


ThreeSum.java
```diff
@@ -0,0 +1,51 @@
+package S30_Codes.Two_Pointers_1;
+
+// Time Complexity : O(n) // n = nums.length
```
daredevilteja, line 3: TC is O(n^2) and SC is O(n)

daredevilteja:
Good Work! Keep it up

PR #1090, opened 2022-10-24
Two Pointers-1: Please review

pavansai1:


p2.java
```diff
@@ -0,0 +1,45 @@
+// TC: O(nlogn)
```
daredevilteja, line 1: TC is wrong

p2.java
```diff
@@ -0,0 +1,45 @@
...
+// SC: O(1)
```
daredevilteja, line 2: SC is wrong

daredevilteja:
Good Work! Keep it up

PR #1094, opened 2022-10-27
Two Pointers

18ksangeeta:


3Sum.py
```diff
@@ -0,0 +1,23 @@
+class Solution:
+    # leetcode URL : https://leetcode.com/problems/3sum/
+    # Problem : 15. 3Sum
+    # Time Complexity : O(n^2)
+    # Space Complexity : O(1)
```
daredevilteja, line 5: SC is O(log n to n) due to sorting

daredevilteja:
Good Work! Keep it up

PR #1100, opened 2022-11-04
Finished-Please review

ddhorajiya:


3sum
```diff
@@ -0,0 +1,33 @@
+//TC : O(n*2)
+//SC: O(1)
```
daredevilteja, line 2: SC is wrong. Please recheck it.

daredevilteja:
Good Work! Keep it up

PR #1122, opened 2022-12-09
Completed Two Pointers - 1 All problems

vigviswa:
Completed all problems, please review

### DIFF
--- a/3sum.py
+++ b/3sum.py
@@ -1,0 +1,38 @@
+# Time Complexity: O(nlogn) + O(n^2) ~ O(n ^ 2)
+# Space Complexity: O(1)
+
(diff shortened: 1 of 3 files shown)


3sum.py
```diff
@@ -0,0 +1,38 @@
+# Time Complexity: O(nlogn) + O(n^2) ~ O(n ^ 2)
+# Space Complexity: O(1)
```
daredevilteja, line 2: Sorting takes up space

daredevilteja:
Good Work! Keep it up.

3sum.py
vigviswa: Understood. Thank you!

PR #1165, opened 2023-02-10
Completed Two Pointers 1. Please review.

karandhodi:


Problem_2.java
```diff
@@ -0,0 +1,49 @@
+// Time Complexity : O(n^2logn)
```
osandeepreddy, line 1: TC=O(n^2)

osandeepreddy:
TC=O(n^2)

PR #1172, opened 2023-02-13
finished two pointers1 ---- please review

summerhanyuezheng:


p2.py
```diff
@@ -0,0 +1,28 @@
+
+#successfully submitted on Leetcode: yes
+#TC = O(nlogn + n^2)
+#SC = O(1)
```
daredevilteja, line 4: recheck sc

daredevilteja:
Good Work! Keep it up.

PR #1178, opened 2023-03-14
Completed all Problems

shivangishakya:


Problem_36.py
```diff
@@ -0,0 +1,32 @@
+# TC: O(n^2)
+# SC: Worst Case: O(n)
```
osandeepreddy, line 2: SC is O(1)

osandeepreddy:
SC is O(1)

PR #1180, opened 2023-03-14
Completed Sort colors. pls review it. 

sivakami-projects:


pvk3081 (commented):
Please add other two problems also

pvk3081:
Please add other two problems also

PR #1183, opened 2023-03-16
Completed Two-Pointers-1. Please Review.

Tirthaj:


Rohan-Balakrishna:
Good Job Tirthaj...To Answer your Question any output that can be type casted by python to display output can allowed as return object. Its been mention lists of lists to avoid this confusion of castability.

Rohan-Balakrishna:
Good Job Tirthaj...To Answer your Question any output that can be type casted by python to display output can allowed as return object. Its been mention lists of lists to avoid this confusion of castability.

PR #1212, opened 2023-06-05
Completed two Pointers 1- Please review

RohitSaha-RN79MAY2023:


ThreeSum.java
```diff
@@ -0,0 +1,32 @@
+import java.util.*;
+
+public class ThreeSum {
+    //time complexity" for sorting its n(log n) // not sure of the other part
```
ishika2741, line 4: Kindly write the approach that you are using first. 
For two pointers approach + sorting => TC = O(nlog n) + O(n^2) which is the optimised one and you don't require any extra space here as well because you will be moving your pointer till the time you are not finding the unique element , so you don't require to handle duplicate pairs.
for BS + sorting => TC = O(n^2 log n)

ThreeSum.java
```diff
@@ -0,0 +1,32 @@
...
+    // space complexity: constant
```
ishika2741, line 5: In your solution , you used a set , that would require some space.

ishika2741 (commented):
Good work! For problems like three sum , you should know and submit all the solutions(hashing , two pointers and BS) .

ishika2741:
Good work! For problems like three sum , you should know and submit all the solutions(hashing , two pointers and BS) .

PR #1224, opened 2023-06-08
Completed Two-Pointers-1

pkhanna88:


Problem_2.py
```diff
@@ -0,0 +1,35 @@
+# Time Complexity: O(n^2)
+# Space Complexity: O(ans)
```
ishika2741, line 2: ans is the output and output is never considered in complexity.  SC = O(1)

Problem_3.py
```diff
@@ -0,0 +1,22 @@
+# Time Complexity: O(1)
```
ishika2741, line 1: Kindly check TC

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1265, opened 2023-07-17
Completed Two-Pointers-1

PurnaSahithiAdduri:
Completed Two-Pointers-1

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,59 @@
+// Time Complexity : O(nlogn+n^2) ~= O(n^2)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode :yes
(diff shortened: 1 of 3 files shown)


Problem2.java
```diff
@@ -0,0 +1,59 @@
+// Time Complexity : O(nlogn+n^2) ~= O(n^2)
```
ishika2741, line 1: Submit two pointers solution as well.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

Problem2.java
PurnaSahithiAdduri: Added a line 34 of this file

PR #1284, opened 2023-08-18
Two-Pointers-1

deeksha9107:


Exercise_2.py
```diff
@@ -0,0 +1,55 @@
+# ## Problem2 (https://leetcode.com/problems/3sum/)
+# // Time Complexity :O(nlogn)
```
ishika2741, line 2: TC will be O(n^2)

ishika2741 (commented):
Good work Deeksha!

ishika2741:
Good work Deeksha!

PR #1330, opened 2023-10-21
Two-Pointers-1 Problems 2 & 3 - Done

Mohini-Mayekar:
Problem#35 -  3 sum
Problem#36 -  Container With Most Water

### DIFF
--- a/ThreeSum.java
+++ b/ThreeSum.java
@@ -1,0 +1,45 @@
+/* Time Complexity : O(n^3)
+ * 	n - length of the input array - height */
+/* Space Complexity : O(k)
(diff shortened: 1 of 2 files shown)


ThreeSum.java
```diff
@@ -0,0 +1,45 @@
+/* Time Complexity : O(n^3)
```
ishika2741, line 1: Try to submit optimized solution using two pointers

ishika2741 (commented):
Good work! Kindly submit third problem as well.

ishika2741:
Good work! Kindly submit third problem as well.

PR #1366, opened 2024-01-05
Completed two-pointer-1

keerthana-vijaykumar:


Problem-2.cpp
```diff
@@ -0,0 +1,51 @@
+// Time Complexity : O(nlogn)
```
ishika2741, line 1: Check TC

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1417, opened 2024-03-15
Two-Pointer-1 completed

sreejareddy17:


Nandha2Tech (commented):
Great Work! For 3Sum problem TC would be O(n log n) for sorting and another O(n^2) for finding the sum of the numbers that results to target. Since n^2 is the bigger term here TC would be O(n^2).

Nandha2Tech:
Great Work! For 3Sum problem TC would be O(n log n) for sorting and another O(n^2) for finding the sum of the numbers that results to target. Since n^2 is the bigger term here TC would be O(n^2).

PR #1478, opened 2024-07-06
completed two-pointer-1

dhikki-study:
##########11. Container With Most Water####################################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : No

// Your code here along with comments explaining your approach in three sentences only
Took left and right pointer, than found the area, now we have to move either left or right pointer we decided by comparing the height at each pointer which ever is the smallest we will move the pointer ahead and check new area and compare that area to initially calculate area. We will keep moving pointer till left crosses right	
        

#########75. Sort Colors##############################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I sorted the color using extra space but after lecture learnt way to sort color without using extra space

// Your code here along with comments explaining your approach in three sentences only
We have 3 pointers left, right and mid. Left will help to collect 0 mid will help to collect 1 and right will collect zero. Here mid will iterate through the loop and check if it has 0 or 2 if it has 0 than it will replace with left if 2 than replace with right or if it has 1 it will keep moving
        

#########15. 3Sum###################################################
// Time Complexity : n^2
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I was not able to able to extra while check to be applied, I have highlighted same with *** due to which my code was going out of range. After checking the video was able to figure it out

// Your code here along with comments explaining your approach in three sentences only
We have fixed 1 element and applied 2sum on rest of elements. Firstly 0th index element is fixed and 2 sum is applied on rest than next looping through each index and applying 2sum for all other

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,114 @@
+##########11. Container With Most Water#############################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+// Your code here along with comments explaining your approach in three sentences only
+Took left and right pointer, than found the area, now we have to move either left or right pointer we decided by comparing the height at each pointer which ever is the smallest we will move the pointer ahead and check new area and compare that area to initially calculate area. We will keep moving pointer till left crosses right	
+
+
+class Solution:
+    def maxArea(self, height: List[int]) -> int:
+        l,r=0,len(height)-1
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1619, opened 2025-02-06
done

S565730:


Erlybird:
Please optimize 2nd and 3rd problems. Please mention the TC and SC of each solution

Erlybird:
Please optimize 2nd and 3rd problems. Please mention the TC and SC of each solution

PR #1650, opened 2025-02-18
Done Two-Pointers-1

pavan0926:


3Sum.java
```diff
@@ -0,0 +1,26 @@
+// Time Complexity : O(n2)
```
ishika2741, line 1: Please submit the optimal solution using two pointers

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1729, opened 2025-06-15
Completed

solankiram2023:


super30admin:
- Strengths: The solution is correct and easy to understand. The use of counting is a good approach for this problem.
- Areas for Improvement: The solution could be optimized to use a single pass instead of two. The reference solution demonstrates how to do this with two pointers, which is more efficient. Additionally, the space usage could be further optimized by avoiding the counts array.

PR #1825, opened 2025-12-23
Solved two pointers

aahiawasthee:


super30admin:
- Strengths:
  - Correct implementation of all three problems.
  - Good handling of edge cases, especially in "3-sum" with duplicate checks.
  - Clear and readable code structure.
- Areas for Improvement:
  - The time complexity for "container-max-water" is incorrectly noted as O(logn); it should be O(n) since it's a two-pointer approach that scans the array once.
  - In the "sort-colors" solution, there's a typo in the swap method parameter name ("loh" instead of "low" or "high"), though it doesn't affect functionality.
  - While not incorrect, the "3-sum" solution could benefit from a brief comment explaining the overall approach for better readability.

PR #1850, opened 2026-02-21
[sort colors && 3 sum && max container]

takchiks:


super30admin:
Your solution for sortColors is excellent. It efficiently sorts the array in one pass with constant space, meeting all the problem requirements. The code is clear and well-commented.

However, note that for a specific problem, you should only submit the code relevant to that problem. The inclusion of threeSum and maxArea methods is unnecessary and might cause confusion. In a real coding interview or on a platform like Leetcode, you should only provide the solution for the problem at hand.

One minor suggestion: Consider using a helper method for swapping to make the code more modular and avoid duplication. For example:
```java
private void swap(int[] nums, int i, int j) {
    int temp = nums[i];
    nums[i] = nums[j];
    nums[j] = temp;
}
```
Then you can call this method whenever you need to swap elements. This improves readability and reduces code duplication.

Overall, your sortColors method is correct and efficient. Keep up the good work!

PR #1851, opened 2026-02-21
Completed Two-Pointers-1-3Sum

sarvanibaru:


super30admin:
It appears that you have submitted a solution for the "3Sum" problem instead of the "Arrange Colors" problem. Please double-check the problem statement and ensure you are solving the correct problem.

For the "Arrange Colors" problem, you need to sort an array containing only 0s, 1s, and 2s in-place without using the library sort function. The recommended approach is to use the Dutch National Flag algorithm, which uses three pointers: low, mid, and high. The idea is to partition the array into three sections: 0s at the beginning, 1s in the middle, and 2s at the end.

Here is a brief explanation of the algorithm:
- Initialize low = 0, mid = 0, high = nums.length - 1.
- While mid <= high:
  - If nums[mid] is 0, swap it with nums[low], then increment both low and mid.
  - If nums[mid] is 1, just increment mid.
  - If nums[mid] is 2, swap it with nums[high], then decrement high (but do not increment mid because the swapped element from high needs to be checked).

This algorithm runs in O(n) time and uses O(1) space.

Please implement this algorithm for the correct problem. Your current solution for 3Sum is good for that problem, but it is not applicable here.

PR #1853, opened 2026-02-21
Completed Two-Pointers-1-ContainerWithMostWater

sarvanibaru:


super30admin:
It seems there has been a mix-up. The solution you provided is for the "Container With Most Water" problem, not the "Arrange Colors" problem. For the "Arrange Colors" problem, you need to sort an array containing only 0s, 1s, and 2s in-place without using the library's sort function.

To solve the "Arrange Colors" problem, you can use the Dutch National Flag algorithm, which uses three pointers (low, mid, high) to partition the array into three sections: 0s, 1s, and 2s. The algorithm works in one pass with O(n) time complexity and O(1) space complexity.

Here is a brief explanation of the approach:
- Initialize three pointers: low at the start, high at the end, and mid at the start.
- While mid <= high:
  - If nums[mid] is 0, swap it with nums[low], then increment both low and mid.
  - If nums[mid] is 1, just increment mid.
  - If nums[mid] is 2, swap it with nums[high], then decrement high (but do not increment mid because the swapped element from high needs to be checked).

You should implement this algorithm for the given problem. Please ensure you are solving the correct problem in the future.

PR #1855, opened 2026-02-22
Working solution

avcode3:


super30admin:
Your solution for the "Arrange Colors" problem is well-implemented and efficient. You correctly used the three-pointer approach (Dutch National Flag algorithm) to sort the array in one pass with constant space. The code is clean and easy to understand.

One minor point: in the condition for swapping 0, you swap and then increment both start_ptr and mid_ptr. This is correct because the element that was at start_ptr (which you swap with) must be a 1 (since all 0s have been moved to the left and all 2s to the right), so after swapping you can safely move mid_ptr forward. Similarly, when you see a 2, you swap with end_ptr and decrement end_ptr without moving mid_ptr because the element swapped from the end might be 0, 1, or 2 and needs to be rechecked.

Your solution passes all the test cases and meets the problem constraints. Well done!

